PR #91, opened 2015-02-27
Adding maximum files size limit to keep users from uploading files that ...

cam156:
...will break the system, or timeout before they can be downloaded.

Users can upload 10GB files to our PSU box location. 7GB takes about 45 minutes to download the url expires in 1 hour, the url will expire before a user could download 1 or multiple 10GB files.

This disables the files if they are too big.

screen shot 2015-02-27 at 3 29 00 pm


mbklein:
:+1:

PR #105, opened 2015-10-05
Use checkbox to select all files

awead:
This adds support for selecting all files in a directory. Selecting a checkbox will descend into each subfolder, loading it, and selecting its files.
select-all

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,3 +9,5 @@ require 'rspec/core/rake_task'
 require 'engine_cart/rake_task'
 
 task :default => [:ci]
+require 'jasmine'
+load 'jasmine/tasks/jasmine.rake'
--- a/app/assets/javascripts/browse_everything/behavior.js.coffee
+++ b/app/assets/javascripts/browse_everything/behavior.js.coffee
@@ -37,6 +37,81 @@ $ ->
     $('input.ev-url').each () ->
       $("*[data-ev-location='#{$(this).val()}']").addClass('ev-selected')
 
+  fileIsSelected = (row) ->
+    result = false
+    $('input.ev-url').each () ->
+      if this.value == $(row).data('ev-location')
+        result = true
+    return result
+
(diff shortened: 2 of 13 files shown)


jcoyne:
Looks good, but it would look even better with tests.

awead:
I hear @cam156 is really good with JS tests...

mbklein:
I tried to write a comprehensive test suite for browse-everything when we first wrote it, but I couldn't figure out how to make Jasmine work. :confused:

awead:
I've got Jasmine working...sort of. Now I just need to figure out how to write tests...

awead:
@jcoyne I've added Jasmine to the test suite, however as @mbklein notes, there's a lot of difficulty with getting it to work properly. In order to make the JS testable, I had to make changes to the layout of how the functions were defined. You can see that in the last commit where I have to call a function:

https://github.com/projecthydra-labs/browse-everything/commit/cee9d95f946999aa105106c273e2bf3f7b84e54c#diff-0a43a9fbdccde442ca9a8005c537e169R293

I'm not sure this is the best way to do this, and it would entail a complete rewrite of the entire JS, which is out of scope for this PR. I made these changes because we need them to fulfill at feature request on one our applications, psu-stewardship/archivesphere#27.

At this point, I'd just like to get this merged so we can cut a new release I can move on with my work. We do need to this gem's javascript into a more testable state, and I think this PR takes an initial stab at that, but it should be continued in a separate effort, and ideally by someone (other than me) who's more fluent in the language.

jcoyne:
@awead I didn't demand a Jasmine test. How about a view test for your changes there? Have you considered a feature spec?

awead:
@jcoyne I tried doing a feature test via poltergeist, without success.

awead:
@jcoyne I put in a view test for the added table cell that has the checkbox.

jcoyne:
:+1:

PR #119, opened 2016-04-04
Fix assets

cbeer:


mjgiarlo:
:+1:

jkeck:
:+1: This should prob. require a minor version bump (since we're still pre 1.0).

cbeer:
Agreed; released as v0.10.

PR #121, opened 2016-04-11
declare $.fn.browseEverything() outside of DOM ready

jcoyne:
This allows `$(...).browseEverything()` to be called in any DOM ready block, even if that block gets called before the one in this gem.


jcoyne:
This is not quite ready yet, please don't merge.

jcoyne:
I'm abandoning this, as it needs *major* work to make it work sensibly.

PR #127, opened 2016-07-12
Support turbolinks 5 and previous versions

awead:


app/assets/javascripts/browse_everything/behavior.js.coffee
```diff
@@ -304,7 +304,11 @@ auto_toggle = ->
   triggers = $('*[data-toggle=browse-everything]')
   triggers.each () -> $(this).browseEverything($(this).data())
 
-if Turbolinks?
-  $(document).on 'page:change', auto_toggle
+if Turbolinks && Turbolinks.supported
```
cbeer: How does `Turbolinks?` and `Turbolinks` compile differently?

awead: Don't know. I was just copying what you did in Blacklight.

cbeer: Here's what we have in Blacklight:

```
if (typeof Turbolinks !== 'undefined' && Turbolinks.supported) {
```

I'm just not sure if `if Turbolinks` compiles down to that?

awead: Ah, ok. I was looking at an earlier commit. I'll update this.

PR #129, opened 2016-08-05
Don't wait for auto-toggle if the explicit browseEverything() call is seen first.

mbklein:
In certain situations, the `$(obj).browseEverything()` call might get triggered before the auto-toggled initialization caused by `<button data-toggle="browse-everything"... />`. If that happens, the promise object returned by the `browseEverything()` call gets wiped out later during auto-toggle initialization. This change does two things:

1. Causes manual init to include the auto-toggle parameters _if the target hasn't been initialized yet_.
2. Skips auto-toggle init _for any target that's already been manually initialized_.

IOW, whichever init happens first will handle the toggle parameters, and auto-toggle will only happen if it needs to.

### DIFF
--- a/app/assets/javascripts/browse_everything/behavior.js.coffee
+++ b/app/assets/javascripts/browse_everything/behavior.js.coffee
@@ -183,6 +183,7 @@ $ ->
 
   $.fn.browseEverything = (options) ->
     ctx = $(this).data('context')
+    options = $(this).data() unless (ctx? or options?)
     if options?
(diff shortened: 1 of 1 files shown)


cam156:
@mbklein Looks like it should handle the issue we were seeing. @awead any comments?

awead:
LGTM. It'd be nice if we could have some tests, but that's another matter.

awead:
@mbklein @cam156 let me try it out in Sufia and I'll merge if all's well.

Issue #131, opened 2016-09-08
Update to google-api-client 0.9.x


jcoyne:
This is required for use with Rails 5, because google-api-client 0.8.7 was released which locks to `activesupport < 5`

Upgrade instructions are here: https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md

atz:
Resolved by #134

emanuil-tolev:
Not having this released blocks using sufia 7.2.0 with Rails 5, so would be very interested to see a new release esp. considering the code's in already :). This is a new instance so I'm keen to see if it can be upgraded before we add it loads of user data.

I'm not sure I can help that much with testing it though so it may be easier for whoever has already done it in the past to do it... The bit I can't see how to do is how do I get "sufia ~> 7.2.0", but with browse-everything "tip of master" rather than sufia's usual dependency "browse-everything ~> 0.10.3"?

PR #134, opened 2016-09-08
Update google-api-client to 0.9

atz:
This is required for rails 5+ support, because `0.8` series of the gem are limited to `activesupport` <5.  In the `0.9` series, the dependency on `activesupport` is removed entirely, freeing downstream consumers to use whatever version they wish.


atz:
I obviously don't care about -0.08% coverage drop from removing code.

atz:
This would have been useful to discover earlier, but it is informative:
https://developers.google.com/drive/v3/web/migration

atz:
So I got it to this point:

screen shot 2016-09-10 at 1 22 20 am


With populated date field.

Outstanding issues:

- For Google Drive, the AJAX/treeview does not update when filename or arrow/toggle is clicked.  The query completes fine on the back end and returns results, but despite displaying "100%" the browser does not render them.
- No back or "up" navigation, no way to exit a folder

atz:
I dislike the model of returning HTML partials as their own HTTP resource.  We should be returning JSON and rendering the data on the client side into the table form.

atz:
OK, got listings fully functional.  Only final retrieval outstanding.

screen shot 2016-09-27 at 4 06 24 pm

atz:
Note, there is apparently a longstanding bug in `FileEntry` that labels everything a directory, as shows in the screenshots.  Should be fixed here.

mbklein:
:+1: This looks great, @atz. Thanks for putting in all the work. I'm going to merge, but hold off cutting a release until I* get a chance to do a manual test with a couple of the other providers.

* or anyone

emanuil-tolev:
This blocks using sufia 7.2.0 with Rails 5, so would be very interested to see a new release esp. considering the code's in already :). This is a new instance so I'm keen to see if it can be upgraded 

I'm not sure I can help that much with testing it though so it may be easier for whoever has already done it in the past to do it... The bit I can't see how to do is how do I get "sufia ~> 7.2.0", but with browse-everything "tip of master" rather than sufia's usual dependency "browse-everything ~> 0.10.3"?

emanuil-tolev:
Sorry, should have commented on related issue - deleted to move conversation to #131 .

scherztc:
@atz Were you ever able to retrieve a file from google drive using the 0.9.x google-api-client?

PR #148, opened 2017-02-23
Guard against cross-site scripting attacks

awead:
I'm not sure if this is entirely necessary, or even if it might break things. If your base application is already doing this, what happens if you include an engine that does as well?

Someone notified us that not including this was a security flaw, which is technically true I think, but only if you're using this gem alone and not within the context of another Rails application which (ostensively) would be guarding against these attacks.

Thoughts? @jkeck @jcoyne @jeremyf

### DIFF
--- a/app/controllers/browse_everything_controller.rb
+++ b/app/controllers/browse_everything_controller.rb
@@ -4,6 +4,8 @@ class BrowseEverythingController < ActionController::Base
   layout 'browse_everything'
   helper BrowseEverythingHelper
 
+  protect_from_forgery with: :exception
+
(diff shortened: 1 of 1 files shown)


cbeer:
I wonder why we extend `ActionController::Base` instead of `ApplicationController` (which would give downstream applications the ability to control things like this)?

jeremyf:
@awead Looks good. I approve, but am curious about pursuing @cbeer's suggestion as well. @cbeer care to create a ticket?

jcoyne:
Won't this change prevent json clients from using this endpoint?

PR #149, opened 2017-02-27
Update rubocop and enable for Travis build

awead:
Pins version to 0.42.x which aligns with the current Sufia versions for rubocop and rubocop-rspec.

rubocop_todo was updated using --auto-gen-config and can serve as a starting point for future refactorings.

ping @cam156 this is a prelude to psu-stewardship/scholarsphere#598


cam156:
@awead do the two commits really need to be separate?  Or is this an instance where squashing them would make sense?

awead:
@cam156 yeah, I'm going to squash them... I'm still working on it. Let me make a note.

PR #150, opened 2017-02-28
Enable checkbox selection for files

awead:
Allows you to select a file by using a checkbox, the same way you would select a directory of files.

### DIFF
--- a/app/assets/javascripts/browse_everything/behavior.js.coffee
+++ b/app/assets/javascripts/browse_everything/behavior.js.coffee
@@ -7,7 +7,7 @@ $ ->
 
     dialog.modal
       backdrop: 'static'
-      show:     false 
+      show:     false
     ctx =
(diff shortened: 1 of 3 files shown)


awead:
ping @cam156 I'll post some screenshots once S3 is fully recovered.

app/assets/javascripts/browse_everything/behavior.js.coffee
```diff
@@ -100,9 +104,11 @@ $ ->
           else
             if action == "1"
               $(this).addClass('ev-selected')
+              $(this).find('.ev-select-file').prop('checked', true)
```
cam156: If this code was in the selectFile method I think it would only need to be in one place, not in both.

awead:
@cam156 changes applied. If you merge this, I have a follow-up PR to fix this issue with truncating directory names.

PR #151, opened 2017-03-01
Set size of container columns

awead:
This allows directories with long names to properly truncate the same way as files.

Additionally, the scss and view code was cleaned up for better readability.


awead:
Before:
before
After:
after

PR #153, opened 2017-03-10
Refactoring Driver::S3 to reduce code complexity

awead:
@cam156 from today's refactoring session. We whittled down our ACB size and now have one less overly-long method.


lib/browse_everything/driver/s3.rb
```diff
@@ -6,6 +6,8 @@ class S3 < Base
       DEFAULTS = { signed_url: true, region: 'us-east-1' }.freeze
       CONFIG_KEYS = [:app_key, :app_secret, :bucket].freeze
 
+      attr_reader :file_entries
```
cam156: @awead the name file_entries is indicative of just files being listed to me should it be file_and_path_entries, or directory_listing_entires?

lib/browse_everything/driver/s3.rb
```diff
@@ -20,23 +22,35 @@ def validate_config
         raise BrowseEverything::InitializationError, "Amazon S3 driver requires #{CONFIG_KEYS.join(',')}"
       end
 
+      # @return [Array<BrowseEverything::FileEntry>]
```
cam156: Yeah!  Yard docs!!! 👏

lib/browse_everything/driver/s3.rb
awead: or bucket_entries, but yes, something that means "everything that is in this path"

awead:
@cam156 I named it `file_entries` because they were all instances of BE::FileEntry. However, a FileEntry can be a file or a directory. So you make a good point! I renamed it to just "entries" to keep it inline with the class.

PR #155, opened 2017-03-16
address rubocop issues in filesystem driver

derekmorr:
This addresses the rubocop issues in the filesystem driver.  Specifically, the ABC size, cyclomatic complexity, method length, and rescue modifier issues.


lib/browse_everything/driver/file_system.rb
```diff
@@ -6,24 +6,32 @@ def icon
...
+        entries = if File.directory?(real_path)
+                    make_directory_entry(relative_path, real_path)
+                  else
+                    [details(real_path)]
+                  end
+
+        sort_entries(entries)
+      end
+
+      def make_directory_entry(relative_path, real_path)
```
jcoyne: Can you make these all private methods?

lib/browse_everything/driver/file_system.rb
```diff
@@ -36,18 +44,32 @@ def details(path, display = nil)
...
           info.mtime,
           info.directory?
         )
       end
 
+      def make_pathname(path)
+        Pathname.new(File.expand_path(path)).relative_path_from(Pathname.new(config[:home]))
+      end
+
+      def get_file_size(path)
```
jcoyne: `file_size` would be a more idiomatic name.

lib/browse_everything/driver/file_system.rb
```diff
@@ -36,18 +44,32 @@ def details(path, display = nil)
...
+        File.size(path).to_i
+      rescue
```
jcoyne: I don't you suppose you know what exception the original author was checking for do you?

lib/browse_everything/driver/file_system.rb
```diff
@@ -36,18 +44,32 @@ def details(path, display = nil)
         return nil unless File.exist?(path)
         info = File::Stat.new(path)
         BrowseEverything::FileEntry.new(
-          Pathname.new(File.expand_path(path)).relative_path_from(Pathname.new(config[:home])),
+          make_pathname(path),
           [key, path].join(':'),
-          display || File.basename(path),
+          make_display(path, display),
```
jcoyne: Rather than this, how about changing the signature to `details(path, display = File.basename(path))`? Then you could just use `display` here.

jcoyne:
@derekmorr These changes are great. I don't suppose you have a CLA on file with us, do you?

derekmorr:
Yes, I do. See https://github.com/projecthydra/browse-everything/pull/154#issuecomment-286805099.

I'll address your comments and push up a new commit.

lib/browse_everything/driver/file_system.rb
derekmorr: It looks like `File.size()` throws `Errno::ENOENT` on non-existent files, and returns 0 for files it can't open, or special files (entries under `/dev`, for example).

PR #161, opened 2017-04-25
Add --skip-listen for Rails 5-based builds to avoid fsevent problems

jcoyne:


awead:
@jcoyne I'm good with this despite coveralls, unless you want to tweak it more.

jcoyne:
@awead I'm done. Coveralls shows that no files have changed coverage, so I don't know what's up with that metric.

PR #165, opened 2017-05-05
Stop explicitly requiring jquery

### LINKED ISSUE #163: including browse_everything.js includes a second copy of jquery in the application.js
https://github.com/projecthydra/browse-everything/blob/3c3f60472c2b31a3d5a07e8f2bf50f09f8220ed5/app/assets/javascripts/browse_everything.js#L1

This makes it impossible to use jquery2 or jquery3

It might be better to expect the host application to already have included jquery as we are doing in hyrax: https://github.com/projecthydra-labs/hyrax/pull/891/files

jcoyne:
Instead we'll expect the host app to provide this dependency.
This allows the host app to switch jquery versions (from 1 to 2 or 3)

Fixes #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ download the files.
 
 ## Installation
 
-Add this line to your application's Gemfile:
+Add this lines to your application's Gemfile:
 
+    gem 'jquery-rails'
     gem 'browse-everything'
 
 And then execute:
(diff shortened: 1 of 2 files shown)


cbeer:
Should there also be a documentation or generator update to tell downstream applications they need jquery (especially since Rails 5.1 got rid of it by default)?

PR #167, opened 2017-05-08
Fix typo in Retriever and add test coverage

mjgiarlo:


lib/browse_everything/retriever.rb
```diff
@@ -53,7 +53,7 @@ def retrieve(spec)
           yield(chunk, retrieved, spec['file_size'].to_i)
         end
       else
-        raise URI::BadURIError, "Unknown URI scheme: #{uri.scheme}"
```
mjgiarlo: `uri` here raises a `NameError` because it doesn't exist. `url` is used in this method.

PR #171, opened 2017-05-31
Remove Skydrive support

jcoyne:
Skydrive has become onedrive.  Unclear if this is still useful, so
removing it. The skydrive gem and its dependencies are unmaintained.


jcoyne:
@mbklein Can you have a look?

PR #173, opened 2017-07-05
Long filenames

awead:
This fixes a problem documented in https://github.com/psu-stewardship/scholarsphere/issues/834

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,3 +38,7 @@ else
   end
 end
 # END ENGINE_CART BLOCK
+
+unless File.exist?(file)
+  eval_gemfile File.expand_path('spec/test_app_templates/Gemfile.extra', File.dirname(__FILE__))
+end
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -38,3 +38,7 @@ else
   end
 end
 # END ENGINE_CART BLOCK
+
+unless File.exist?(file)
+  eval_gemfile File.expand_path('spec/test_app_templates/Gemfile.extra', File.dirname(__FILE__))
```
jcoyne: Can we inject the gem via the test app generator like this? https://github.com/projectblacklight/blacklight/blob/610310f89e0763d7632f861328c3828139415bc3/lib/generators/blacklight/assets_generator.rb#L29-L31

awead: @jcoyne sure. Will `bundle` get re-run or do I have to call that somehow? BTW, I was following the same pattern in Hyrax, but maybe this one here is more current?

jcoyne: I was trying to enforce consistency, but it looks like I was the one to break consistency. 🙄 https://github.com/samvera/hyrax/commit/2867afaff64180fc80df7bdd39a2a364cb356185
Let's just :shipit:

Issue #177, opened 2017-08-24
Google Drive: link_for method returns empty url


scherztc:
Since moving up to Google Drive Api version 3 the link_for method is returns an empty url.

This empty url does not let browse_everything retrieve the file.

scherztc:
@mjgiarlo Attaching a file from google_drive is broken.  The attached file display only shows the name.  It displays 0 KB and web_content_link is Nil.  Creating empty records in our hyrax application.

Our engine_cart test application does not increment the attached file display when using google_drive.

Currently there are no unit test for google_drive : https://github.com/samvera/browse-everything/tree/master/spec/unit.  I can certainly add one when I figure out what is going on.

awead:
@scherztc You can write a unit test for Google Drive by cribbing from one of the existing tests for the other drives. Take a look at what I did for https://github.com/samvera/browse-everything/blob/master/spec/unit/box_spec.rb, in particular, https://github.com/samvera/browse-everything/blob/master/spec/unit/box_spec.rb#L16-L28

I had to have a live-running connection during the test to see what was actually coming back from Box, then write the tests. Once I the tests passed with the live data from Box, I recorded the responses with VCR. The annoying thing with that was that the tokens expired after an hour.

scherztc:
https://github.com/samvera/hyrax/blob/77140b38907158aaa4e30aae442ea5f04d0b1be2/app/jobs/import_url_job.rb#L53

I feel like URI isn't enough for Google Drive to work using the retriever. I think we need more header information, particularly mime-type. HTTPParty is returning a Moved Temporarily error.

mjgiarlo:
How is this looking, @jrgriffiniii?

jrgriffiniii:
@mjgiarlo a rudimentary unit test has been provided using a [WIP] PR, and I'm going to continue looking towards reproducing the empty URL issue.

mjgiarlo:
@jrgriffiniii I'd probably tackle #159 in advance of the Hyrax integration work just in case you find that the new Dropbox API also requires changes to `ImportUrlJob` -- then you can tackle that in one new Hyrax PR instead of two. Does that make sense?

jrgriffiniii:
@scherztc returning to the concerns you expressed over `BrowseEverything::Retriever` within `ImportUrlJob` for Hyrax, do you feel as if this is progressing in the right direction:
[...]

PR #178, opened 2017-08-24
Fixes google drive empty web_content_link

scherztc:
Fixes: #177

The drive.get_file call needed a field parameter to get the web_content_link.

file = drive.get_file(id, fields: 'id, name, size, web_content_link')


scherztc:
screen shot 2017-08-28 at 4 57 54 pm

Still getting a weird error message when downloading file and using web_content_link

scherztc:
Might need to back port this 
https://github.com/samvera/hyrax/commit/f156ef24b11c9716165312531807cb8aeb57f2fe

scherztc:
https://github.com/samvera/hyrax/blob/77140b38907158aaa4e30aae442ea5f04d0b1be2/app/jobs/import_url_job.rb#L53

I feel like URI isn't enough for Google Drive to work using the retriever.  I think we need more header information, particularly mime-type.  HTTPParty is returning the above error in my temp file.

mjgiarlo:
@scherztc Why isn't this line automatically following the redirects?

https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/retriever.rb#L75

In HTTParty, `follow_redirects` is set to `true` by default.

mjgiarlo:
@scherztc @jrgriffiniii Rubocop now fails. Would someone mind tossing a commit at this branch to get the build green?

scherztc:
@mjgiarlo I think BE is ignoring the follow_redirects because when Hyrax calls the retriever from the ImportUrlJob, it is only passing the uri and not including any header information.   Just my initial guess.

https://github.com/samvera/hyrax/blob/master/app/jobs/import_url_job.rb#L50-L58

mjgiarlo:
@scherztc Looking through the BE and HTTParty internals, it's still not clear to me why they aren't dealing with this. I don't think they should need anything from Hyrax to do the right thing here.

mjgiarlo:
But in any event, we'll need this branch fixed before we can move on to testing a new BE release in Hyrax.

PR #189, opened 2017-12-14
Google Drive Provider Update and Refactoring

jrgriffiniii:
A number of issues needed to be addressed in order to get this working (from my local environment):

- Storing and retrieving the Google API access token using the `session` came with many limitations
  - This was refactored using Service Objects (I have mixed feelings about this approach)
- Working with the supporting Google Ruby Gems came with a number of complications, but the usage of batch requests is now integrated
- Populating the files into the `_files` template partial was not optimal; this has been changed for the purposes of speed
- Google Drive API method `web_content_link` does *not* provide the preferred method of retrieving the URL best used for downloads (!) . Discovering this while testing was not a quick process.

Documentation needs to be improved upon this (and likely testing as well).


app/services/browse_everything_session/provider_session.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def token=(value)
+      @session["#{@name}_token"] = value
+    end
+
+    def token
+      @session["#{@name}_token"]
+    end
+
+    def code=(value)
```
scherztc: +1  to code= as a method name.

jrgriffiniii: @scherztc this is a Ruby convention for accessor methods.  Normally `attr_writer` or `attr_accessor` or suffice if simple mutation is necessary, but here I was trying to be explicit (and avoid metaprogramming) for the purposes of clarity.

jrgriffiniii:
Test coverage has been increased and this is prepared for another review please.

jrgriffiniii:
@scherztc is there any chance that you might be available to review this please?

lib/browse_everything.rb
```diff
@@ -4,8 +4,6 @@
 require 'browse_everything/retriever'
 
 module BrowseEverything
-  class InitializationError < RuntimeError; end
```
escowles: Does this exception class need to be moved to the helper?  It seems like it might break apps if the exception class is renamed.

jrgriffiniii: This has been moved to the `BrowseEverything` Module:
https://github.com/samvera/browse-everything/pull/189/commits/7ac541e34ccd9a7b91811e503f3ec1253d351372#diff-d9f2a3b3a4c2e755ea28e3e468addbbdR23

PR #198, opened 2018-02-26
Provides remedy for cases where Google Drive auth tokens must be forcibly cleared

jrgriffiniii:
Also modifies the default sorting for file entries returned from the Google Drive API


mjgiarlo:
I believe this can now be rebased and resubmitted. Thanks, @jrgriffiniii

Issue #204, opened 2018-04-05
Ensure that directories containing "." characters are handled properly by the `FileSystem` Driver


jrgriffiniii:
One approach would be to impose some constraints upon the `browse_everything#show` route: https://github.com/samvera/browse-everything/compare/master...jrgriffiniii:hotfix-jrgriffiniii-file-systems-punct

jrgriffiniii:
Rescoping this issue in response to bug(s) discovered by @jrochkind

PR #205, opened 2018-04-06
Ensures that file and directory paths containing "." characters resolve

jrgriffiniii:
Progresses towards resolving #204


config/routes.rb
```diff
@@ -3,6 +3,6 @@
 BrowseEverything::Engine.routes.draw do
   get 'connect', to: 'browse_everything#auth', as: 'connector_response'
   match 'resolve', to: 'browse_everything#resolve', as: 'resolver', via: %i[get post]
-  match ':provider(/*path)', to: 'browse_everything#show', as: 'contents', via: %i[get post]
+  match ':provider(/*path)', to: 'browse_everything#show', as: 'contents', via: %i[get post], format: false
```
awead: A comment here would be useful. It's not immediately evident why no format options should be allowed.

PR #235, opened 2018-08-29
Supports pagination for Google Drive file entries and restricts the scope of the Google API client to read-only access

jrgriffiniii:
Currently, the Google API for Drive assumes that the client paginates through file entries (please see the examples in "Using Client libraries" for https://developers.google.com/drive/api/v3/search-parameters).  This is not supported in `browse-everything`.

This offers to integrate pagination for the Google Drive provider by implementing `BrowseEverything::Driver::Paginator::GoogleDrive`.  There is also a default implementation (which is tested but not used) in `BrowseEverything::Driver::Paginator::Base`.  The latter is meant to establish a pattern by which to support pagination for other providers.


mbklein:
@jrgriffiniii Do you have some time to look into these conflicts and resolve them against master? (You might want to wait until #263 is merged in case it piles on more conflicts)

jrgriffiniii:
Deprecated in favor of offering support for pagination across all providers (please see https://github.com/samvera/browse-everything/issues/268).

PR #236, opened 2018-09-17
Removes VCR from the test suites for the drivers

### LINKED ISSUE #233: Better stubbing of HTTP requests
The VCR gem is great at what it does, but not great at rolling with small changes to requests (e.g., parameter order), so tests tend to break with minor updates to provider API dependencies. We should find a better way to stub these requests, especially where OAuth is involved.

jrgriffiniii:
Resolves #233 by ensuring that the Box, Dropbox, and Google drivers exclusively use WebMock for testing


tpendragon:
@mbklein I approved this, but given this was a request from you would you mind taking a look?

PR #248, opened 2018-11-16
BrowseEverything::ConfigurationError for missing configuration yml

### LINKED ISSUE #247: `Errno::ENOENT` should not be raised when the config. file is missing
Currently this affects Hyrax implementations, and was discovered by @seanupton:
> Yes, default for `Hyrax.config.whitelisted_ingest_dirs` will be `Array.wrap(BrowseEverything.config['file_system'].try(:[], :home)).compact` — AFAICT, means that `BrowseEverything.config['file_system'].try(:[], :home)` can return either `nil` or `[]` to make Hyrax have an empty whitelist by default.
(Please see https://samvera.slack.com/archives/C0F9JQJDQ/p1542394129514900)

bmquinn:
Raise `BrowseEverything::ConfigurationError` if `browse_everything_providers.yml` is missing.

Fixes #247

### DIFF
--- a/lib/browse_everything.rb
+++ b/lib/browse_everything.rb
@@ -40,6 +40,7 @@ module BrowseEverything
   end
 
   class InitializationError < RuntimeError; end
+  class ConfigurationError < StandardError; end
   class NotImplementedError < StandardError; end
   class NotAuthorizedError < StandardError; end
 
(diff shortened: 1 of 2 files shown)


lib/browse_everything.rb
```diff
@@ -40,6 +40,7 @@ module Google
   end
 
   class InitializationError < RuntimeError; end
+  class ConfigurationError < StandardError; end
```
no-reply: I'm tempted to suggest this should inherit `Errno::ENOENT`, instead of `StandardError`.

If I handle the error emitted by current versions, this will break my handling.

PR #251, opened 2018-11-19
Replace all CoffeeScript with JavaScript

mbklein:
This PR removes `app/assets/javascripts/browse_everything/behavior.js.coffee` in favor of the equivalent `.js` file, generated using the [Decaffeinate](https://github.com/decaffeinate/decaffeinate) utility.

I also replaced the jasmine test runner with karma to make the JS tests run more cleanly.


spec/javascripts/karma_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+require 'open3'
+
+# Run karma and parse the output for failures.
+# The spec will fail if any karma test fails.
+describe 'Karma' do
+  let(:runner)  { Open3.capture3('karma', 'start') }
+  let(:output)  { runner[0] + runner[1] }
+  let(:status)  { runner[2].exitstatus }
+
+  it 'expects all karma tests to pass' do
```
jrgriffiniii: 👍, but I'm not certain how this is going to integrate with test coverage analysis in the future.  My hope would be to integrate a tool such as https://github.com/istanbuljs/nyc

PR #265, opened 2019-02-15
Ensures that Dropbox downloads are not placed within a directory outside of the Rails app

jrgriffiniii:
Currently, the Dropbox provider works by temporarily downloading the file from a remote storage server.  This file is created in a temporary location (please see https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/driver/dropbox.rb#L94).

Unfortunately, in order to this to integrate with Hyrax, this directory must be explicitly whitelisted, leading to errors when using Dropbox to upload files for `FileSets`: https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/config/initializers/hyrax.rb#L262

This also provides a configuration option for the Dropbox provider download directory path (as well as fixes the CircleCI config. for building the internal test app. using `engine_cart`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ jobs:
         command: bundle check || bundle install
     - run:
         name: Generate test app
-        command: (git diff --name-only master | grep -qG 'generators\/browse_everything') || bundle exec rake engine_cart:generate
+        command: bundle exec rake engine_cart:generate
     - run:
         name: Ensure test app dependencies are installed
         command: |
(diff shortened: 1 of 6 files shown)


jrgriffiniii:
This passed on CircleCI: https://circleci.com/workflow-run/6c26e400-1b14-4a54-8433-e0f351b774f8 but failed for one job on the Travis CI matrix: https://travis-ci.org/samvera/browse-everything/builds/493778384?utm_source=github_status&utm_medium=notification

jrgriffiniii:
Restarting the individual job ensured that this is passing.  Travis CI jobs might flap due to `bundler` version conflicts in spite of the attempt which was made with #262 to forcibly downgrade `bundler` for all builds.

PR #270, opened 2019-03-13
Support the Hyrax 3.0.0 beta 2 pre-release

jrgriffiniii:
This resolves issues I've discovered when testing against https://github.com/samvera/hyrax/releases/tag/v3.0.0.pre.beta2 using the following providers:
- File System
- Dropbox
- Google Drive
- S3

Following the testing of the Box provider, this should prepare for a 1.0.1 release, and the changes in https://github.com/samvera/hyrax/compare/master...jrgriffiniii:hyrax3-plus-be1.0-jrgriffiniii should be introduced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ lib/bundler/man
 pkg
 rdoc
 spec/reports
+spec/fixtures/Getting Started.pdf
 test/tmp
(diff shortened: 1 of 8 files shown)


mbklein:
:+1:

PR #294, opened 2019-07-31
Tidy up CircleCI / README regarding Ruby/Rails versions

botimer:
Re: samvera/maintenance#20
Re: samvera/maintenance#21


botimer:
There is a little discussion on whether we should even have defaults for the ruby/rails versions. Both values are important and the combination basically defines each of the jobs. We may decide to inline everything.

PR #305, opened 2019-10-21
Introduces support for a `file_system` driver for the 2.0 prototype

jrgriffiniii:
This addresses some bugs discovered while testing during Samvera Connect 2019

PR #307, opened 2019-10-23
Renaming library/lower (i. e. non-REST) API Classes to use `Driver` instead of `Provider`

jrgriffiniii:
This required that I delete the old implementation for Driver support (given the conflict in namespaces for the existing Driver API).  Also, I have a preference for retaining the term `Provider` in the RESTful API, but using `Driver` for interfacing with the actual storage service, but I am definitely open to criticisms of this.

Blocked by #305.


jrgriffiniii:
Merging this in preparation for the first community sprint.

PR #309, opened 2019-10-31
Handle cases where the BrowseEverything configuration is not in place with a warning and empty configuration values

jrgriffiniii:
Please see https://github.com/samvera/hyrax/issues/4140 for context, also **releases version 1.0.2** (this is in order to ensure that the patch release fixes Hyrax 2.5.z and 2.6.z releases)


randalldfloyd:
+1 on inspection and reviewing Slack conversation prompting this PR.

PR #325, opened 2020-02-19
Supporting Rails 5.1.z releases

jrgriffiniii:
This stabilizes the `next/rails-api/jrgriffiniii` branch with support for Rails 5.1.z releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,6 @@ workflows:
           name: "ruby2-7_rails5-1"
           ruby_version: 2.7.0
           rails_version: 5.1.7
-
       - build:
           name: "ruby2-6_rails6-0"
           ruby_version: 2.6.5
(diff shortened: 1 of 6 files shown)


jrgriffiniii:
Somewhere like https://github.com/rails/rails/blob/v5.1.7/activesupport/lib/active_support/number_helper/number_to_rounded_converter.rb#L51 seems to be triggering the following error:

```
ActionView::Template::Error:
       undefined method `new' for BigDecimal:Class
```
in https://circleci.com/gh/jrgriffiniii/browse-everything/1278

This concerns me, and we might not be able to support Rails 5.1.7 on Ruby 2.7.0.

jrgriffiniii:
I'm merging this in order to advance progress with the sprint. This should now clear a path for backporting Ruby/Rails maintenance support, as well as for writing new drivers for 2.0.

Issue #327, opened 2020-02-20
Provide integration for the user interface using Webpack


jrgriffiniii:
This should draw from either the NPM registry or GitHub in order to import React Components or Web Components for the user interface in 2.0.0.

PR #328, opened 2020-02-20
Provides an initial attempt at integrating the React UI with the Rails server for testing

jrgriffiniii:
Advances #327, but blocked by https://github.com/samvera-labs/browse-everything-redux-react/pull/36

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,29 @@ yarn install
 ```
 
 If you wish to use a directory on your local environment, one must instead use
-the following approach:
+one of the following approaches:
+
+##### [yarn link](https://classic.yarnpkg.com/en/docs/cli/link/#toc-yarn-link-in-package-you-want-to-link):
+```bash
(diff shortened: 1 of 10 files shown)


jrgriffiniii:
`peerDependencies` should be investigated for the `package.json`

jrgriffiniii:
I've found that the `yarn link` approach does indeed work for me. However, this seems to break when you use `yarn link "browse-everything-react"` followed by a `yarn install` from within `.internal_test_app`. Unfortunately, `yarn unlink "browse-everything-react"` does not fix these problems, and I needed to freshly clone the local copy of the React GitHub UI in order to get the `yarn link` approach working after past failures. I have attempted to detail a work-around for cases where the build is corrupted in the README.

Hot module reloading works for this also 🎉 , so one should be able to interactively work with the UI now with a stable Rails API running.

Issue #319, opened 2020-02-17
Deprecate server-side browsing of Google Drive files using the Google Picker API


jrgriffiniii:
This follows https://github.com/samvera-labs/browse-everything-redux-react/issues/27, as this would ensure that the Rails server mounting BrowseEverything is no longer involved in mediating asynchronous requests to the Google Drive API.

jrgriffiniii:
I've been advancing this on a branch, and I do have this so that currently one is able to browse the contents of one's own Google Drive:

![image](https://user-images.githubusercontent.com/1443986/75013017-572ef200-5451-11ea-9180-31b3dbc53e9f.png)

There are two possible problems which I've encountered thus far:

`http://localhost:3001` *cannot* be used as a whitelisted origin for the Google Picker API. I have no idea why this is the case, but unless this is somehow related to my own account settings (I don't recall ever having used 3001 on any test applications before, but I may just not be remembering these cases correctly), then we may need to have the UI default to 3333 or something more obscure.

Also, as one can see, the Picker UI is a modal dialogue. I have no strong opinions on this, but the penalty from a design standpoint is there.

jrgriffiniii:
I'm also seeing some 500 errors when requesting thumbnails or metadata:
![image](https://user-images.githubusercontent.com/1443986/75013307-fe138e00-5451-11ea-9e55-688c122914b5.png)

jrgriffiniii:
This is working for me locally with https://github.com/samvera-labs/browse-everything-redux-react/compare/master...jrgriffiniii:issues-319-jrgriffiniii-google-filepicker and https://github.com/samvera/browse-everything/compare/2.x-stable...jrgriffiniii:issues-319-jrgriffiniii-google-picker.

However, there are improvements I would like to introduce to the React UI, along with integration testing I would like to undertake with https://github.com/pulibrary/figgy

PR #331, opened 2020-02-28
Extends the Google Drive support in 2.0.0 by supporting the Google Picker API (also introduces Rails 5.1.z  support)

jrgriffiniii:
This brings in updates which I've introduced in order to support the usage of a Google Picker API within the React user interface (please see the ongoing work on https://github.com/samvera-labs/browse-everything-redux-react/compare/master...jrgriffiniii:issues-319-jrgriffiniii-google-filepicker). This was also developed against an existing Rails 5.1.z app. which is required to use the `0.1` release of `activestorage`.

The Google Picker API is a client-side JavaScript solution for selecting Google Drive items (please see https://developers.google.com/picker/). The interaction with the Rails API for 2.0.0 was most consistent, but there were some adjustments needed for supporting a slightly different approach for OAuth2 authorization.

Advances #319


app/controllers/browse_everything/authorizations_controller.rb
```diff
@@ -2,6 +2,10 @@
 
 module BrowseEverything
   class AuthorizationsController < ActionController::Base
+    include BrowseEverything::Controller::JsonApiRequestable
+
+    skip_before_action :verify_authenticity_token if respond_to?(:verify_authenticity_token)
```
jrgriffiniii: These `respond_to?(:verify_authenticity_token)` were needed in order to support Rails 5.1.z releases.

app/controllers/browse_everything/authorizations_controller.rb
```diff
@@ -22,6 +26,28 @@ def show
...
+      raise ResourceNotFound if authorizations.empty?
+
+      authorization = authorizations.first
+      @authorization = authorization.update(**authorization_attributes)
+      @authorization.save
+
+      @serializer = AuthorizationSerializer.new(@authorization)
+      serialized_json = @serializer.serialized_json
+      serialized = JSON.parse(serialized_json)
+      # I do not know why this is happening
```
jrgriffiniii: This seems to be an issue with the `fast_jsonapi` serialization for `Authorizations`

jrgriffiniii:
I've tested this throughout the past two weeks against a custom Samvera repository for integrating Google Drive uploads. In order to advance a 2.0.0.pre.alpha1 pre-release, I'm going to merge this.

PR #355, opened 2020-06-14
Adding CONTRIBUTING.md

This was uploaded via automation.

jeremyf:


hackartisan (changes requested):
The language section looks good but the other edits don't seem accurate. Perhaps the script that generates this file should be updated?

jrgriffiniii:
This is deprecated by https://github.com/samvera/browse-everything/pull/381

PR #374, opened 2022-01-18
Google drive updates

cjcolvar:
Extracted from work done for Avalon at https://github.com/avalonmediasystem/browse-everything/tree/v0.16.1-gdrive-fixes

I'm going to test out this branch on Avalon.
@jrgriffiniii Do you know of any other apps using the google driver that could be tested with these changes?


cjcolvar:
This was reworked locally and this PR is now out of date.  I'm not sure if I'll be coming back to it since it looks like we'll be using a different adapter due to changes in university agreements.

PR #375, opened 2022-02-11
Speed up file selection by avoiding unnecessary DOM interaction

dlpierce:
Opening and selecting directories containing 1000s of files is slow due to excessive interaction with the DOM. This tracks which files are selected in a map and only appends the resulting input objects when the browse everything dialog is submitted.

Some performance numbers: Previously, simply checking the Select All box on an already expanded file system directory containing 12000 files would take close to 10 minutes to complete. This now completes in about 5 seconds, and then submitting the form takes 30s. (There could be further optimization here.)

### DIFF
--- a/app/assets/javascripts/browse_everything/behavior.js
+++ b/app/assets/javascripts/browse_everything/behavior.js
@@ -2,6 +2,7 @@
 
 $(function () {
   var dialog = $('div#browse-everything');
+  var selected_files = new Map(); // { url: input element object }
 
   var initialize = function initialize(obj, options) {
     if ($('div#browse-everything').length === 0) {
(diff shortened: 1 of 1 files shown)


jrgriffiniii:
I am trying to run this locally, but keep encountering the following errors:

```bash
Looking up gems ["powerbar", "virtus", "power_assert", "needle", "jruby-pageant", "bcrypt_pbkdf", "rbnacl", "rbnacl-libsodium"]
Looking up gems ["pattern-match", "ansi", "hashie", "descendants_tracker", "axiom-types", "coercible", "equalizer"]
Looking up gems ["ice_nine", "adamantium"]
Looking up gems ["memoizable"]
Double checking for ["bundler"] in rubygems repository https://rubygems.org/ or installed locally
Fetching gem metadata from https://rubygems.org/
Looking up gems ["bundler"]
```

jrgriffiniii:
Deprecated by https://github.com/samvera/browse-everything/pull/380

PR #426, opened 2024-10-14
Updating the tool versions and removing Ruby releases prior to 3.1 and Rails releases 6.1

jrgriffiniii:
Updating the tool versions and removing Ruby releases prior to 3.1 and Rails releases 6.1. Ensuring that NodeJS is support is limited to LTS Gallium and bundler 2.5.21


cjcolvar (approved):
Looks good!

PR #446, opened 2025-10-08
Fix Google Drive file listing

JulieWinchester:
The Google Drive gem's DriveService method `list_files` expects keyword arguments ([method link](https://github.com/googleapis/google-api-ruby-client/blob/22662130e99be4630ea9e985976af79aada9056d/generated/google-apis-drive_v3/lib/google/apis/drive_v3/service.rb#L1488)), and while Ruby 2 was fine converting hashes to keywords, Ruby 3 throws an error on doing so. This PR expands the request params hash to work as keyword arguments.


cjcolvar (approved):
👍 We made the same change in our fork.

PR #448, opened 2025-10-09
Make dropbox-api gem an 'optional' dependency callers must add themselves to use dropbox driver

jrochkind:
## Motivation

dropbox-api is fairly unmaintained, and has it's own old dependencies which are causing problems for some browse-everything users that don't even use the dropbox driver.  Including possible security concerns with dependency on unmaintained `oauth2` 1.x version.  Ref #422 

* some people were still using dropbox driver and it was still working for them (Before we realized this, there was a consensus to drop dropbox driver)
* We don't have the current capacity to fix this in a "good" way (fork dropbox-api or rewrite dropbox driver not to use it)

So after some discussion with @cjcolvar and @JulieWinchester , the best thing to do seems to be to drop `dropbox_api` gem from the gemspec, but leave the dropbox driver here -- implementers wanting to use it will have to add `dropbox_api` to their own local Gemspec -- and those who don't want to use the dropbox driver won't have to pay the outdated dependency conflict price. 

Not ideal, but seems like the best solution for now until it can be either improved or dropped. 

## Implementation

The dropbox driver is already loaded with an old-style [traditional ruby stdlib `autoload`](https://github.com/samvera/browse-everything/blob/main/lib/browse_everything.rb#L15B). So it's automatically loaded/interpreted when `BrowseEveryting::Dropbox::Driver` is referenced, but not before. 

[I added a pattern I have used before](https://github.com/samvera/browse-everything/blob/6307ac3af393c395c70b3723dd9cdee8e64d8d3a/lib/browse_everything/driver/dropbox.rb#L7-L13) such that when/if the driver is loaded, it will check for suitable availability of `dropbox_api` dependency, and give you a nice error message if not. Trying to make as good DX as we can for this confusing situation:

dropbox_api is _not_ in gemspec -- but we do need to it to `Gemfile`, so it will be available to run tests, because we need it to pass the tests!  I did remove dropbox as "example" in tests it didn't need to be the example, to decouple further -- but we still gotta test the dropbox driver itself. 

So one downside is we're testing in an environment different from app -- we can't guarantee we don't accidentally have a dropbox dependency elsewhere. 

## Or, would it be simpler?

Should we drop the `autoload`, and say in the new major version of b-e, if you want to use the dropbox driver, you have to explicitly drop `require "browse_everything/driver/dropbox"` into your app boot?  I think that would make things somewhat less confusing and less error prone, and think I would like to do that -- didn't realize until I had analyzed and written this whole thing up (thanks rubber duck). 

Feedback welcome on this approach. Still needs something added to doc to explain it, which is a bit tricky in the old unmaintained docs... or should the note be in the generated config file rather than the README?  (I honestly don't know if the generator currently even works!)

### DIFF
--- a/lib/browse_everything/driver/dropbox.rb
+++ b/lib/browse_everything/driver/dropbox.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
 require 'tmpdir'
-require 'dropbox_api'
 require_relative 'authentication_factory'
+
+# We have an unspecified "optional" dependency, not great but it's what we got
+begin
+  gem 'dropbox_api', '>= 0.1.20'
+rescue Gem::LoadError => e
+  new_raise = Gem::LoadError.new("Loading BrowseEverything::Driver:Dropbox requires the availability of gem dropbox_api '#{e.requirement}', please add it to your Gemfile")
+  new_raise.requirement = e.requirement
+  # raise new_raise, cause: nil # cause is confusing and unneeded here.
+end
 
 module BrowseEverything
   module Driver
--- a/lib/browse_everything/driver/dropbox.rb
+++ b/lib/browse_everything/driver/dropbox.rb
@@ -9,7 +9,7 @@ begin
 rescue Gem::LoadError => e
   new_raise = Gem::LoadError.new("Loading BrowseEverything::Driver:Dropbox requires the availability of gem dropbox_api '#{e.requirement}', please add it to your Gemfile")
   new_raise.requirement = e.requirement
-  # raise new_raise, cause: nil # cause is confusing and unneeded here.
+  raise new_raise, cause: nil # cause is confusing and unneeded here.
 end
 
 module BrowseEverything
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,9 @@ group :development, :test do
   gem 'pry-byebug' unless ENV['CI']
 end
 
+# "soft"/"optional" dependency, but needed for our tests of dropbox driver to pass
+gem 'dropbox_api', '>= 0.1.20'
+
 # == Extra dependencies for dummy test app ==
 #
 # Extra dependencies for dummy test app are in .gemspec as a development dependency
--- a/README.md
+++ b/README.md
@@ -24,6 +24,20 @@ download the files.
 
 **This gem does not depend on hydra-head**
 
+## Certain drivers needing maintenance require extra setup
+
+Certain drivers need additional development work if they are to be fully
+supported again, and for now require extra opt-in setup.
+
+### Dropbox
+
+The driver has a dependency on unmaintained dropbox_api gem with it's [own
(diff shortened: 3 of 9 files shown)


jrochkind:
OK I've dropped the autoload, and doc'd in README the necessity of opt-in dependendency and `require`. 

Feedback welcome!  @cjcolvar @JulieWinchester 

If we like this approach, I'll go on and do similar with `box` and then release a 2.0 pre-release. (Call it alpha I guess, as there could be tons more changes). 

Thank you for any help or confirmation I'm on an acceptable track!

lib/browse_everything/driver/dropbox.rb
```diff
@@ -1,9 +1,17 @@
...
 require_relative 'authentication_factory'
 
+# We have an unspecified "optional" dependency, not great but it's what we got
+begin
+  gem 'dropbox_api', '>= 0.1.20'
+rescue Gem::LoadError => e
+  new_raise = Gem::LoadError.new("Loading BrowseEverything::Driver:Dropbox requires the availability of gem dropbox_api '#{e.requirement}', please add it to your Gemfile")
+  new_raise.requirement = e.requirement
+  # raise new_raise, cause: nil # cause is confusing and unneeded here.
+end
```
cjcolvar, line 13: Is dynamic loading of the gem still necessary if the gem is added to the Gemfile of the implementing app?

cjcolvar: I read the PR description more closely and see this is for better messaging to implementing developers.  That makes sense.

cjcolvar: Looking again does this need to raise somehow?

cjcolvar:
This looks good to me overall!  Thanks for working on this @jrochkind!

lib/browse_everything/driver/dropbox.rb
jrochkind: Thank you!!!! goot catch, I had commented that out for testing things. Thank you for review!

jrochkind: It does not NEED to happen, but what it is is a way to get a very nice error message to the user instead of just `no such class as Dropbox`.  I copied the technique -- and improved it with even better error message -- from what Rails does with "optional" dependencies. 

https://github.com/rails/rails/blob/252ee6c6b846c6ce4a82528d03c2bd6d89c7a0fa/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L6

https://bibwild.wordpress.com/2015/09/09/optional-gem-dependencies/

this is a weird and confusing situation, so I love a good Developer eXperience that actually gives you an error message like:

`Loading BrowseEverything::Driver:Dropbox requires the availability of gem dropbox_api >= 0.1.20, please add it to your Gemfile`

jrochkind: oh, the technique is also nice for letting you specify what range of versions you accept, which is what Rails does -- you can error if what they have in their gemfile/gemfile.lock is not in the range you want to require. (Not really using this here so much, but sets it up for easy possibility).

cjcolvar (approved):
Looks great to me!

PR #454, opened 2025-12-10
Bump version to 2.0.0.alpha.2

jrochkind:
The ONLY difference this release will have from 2.0.0.alpha.1 is allowing Rails 8.1, but let's get that out there before possibly larger changes are merged. 

https://github.com/samvera/browse-everything/compare/v2.0.0.alpha.1..main


cjcolvar (approved):
👍

